PR #7, opened 2025-02-18
refactor doc

gogo2464:
I had also set a easier way to navigate to the doc. I will merge all in hugo as soon as possible.

Issue #3, opened 2024-08-02
Home page not updating


knurdtech:
Paragraph on home page has said "For the past two weeks..." for almost two months. I quickly tried to get something together but didn't set up a test environment to try it.

`<span id="elapsed-time"></span>`
<pre>
<script>
document.addEventListener('DOMContentLoaded', (event) => {
    const previousDate = new Date('2024-07-04'); // Set your previous date here
    const today = new Date();
    
    const elapsedTime = getElapsedTime(previousDate, today);
    document.getElementById('elapsed-time').innerText = elapsedTime;
});

function getElapsedTime(startDate, endDate) {
    const msPerDay = 24 * 60 * 60 * 1000;
    const msPerMonth = 30 * msPerDay;
    const msPerYear = 365 * msPerDay;

    const elapsed = endDate - startDate;

    if (elapsed < msPerMonth) {
        const days = Math.floor(elapsed / msPerDay);
        return `For the past ${days} ${days === 1 ? 'day' : 'days'}`;
    } else if (elapsed < msPerYear) {
        const months = Math.floor(elapsed / msPerMonth);
        return `For the past ${months} ${months === 1 ? 'month' : 'months'}`;
    } else {
        const years = Math.floor(elapsed / msPerYear);
        return `For the past ${years} ${years === 1 ? 'year' : 'years'}`;
    }
}
</script>
</pre>

PR #26, opened 2026-04-12
refactor: rename ghost-key delegate → notary (0.2.0, closes #24)

### LINKED ISSUE #24: Rename ghost-key 'delegate' → 'notary' to deconflict with Freenet delegates
# Rename ghost-key "delegate" → "notary" to deconflict with Freenet delegates

## Context

The `ghostkey_lib` crate (and the `ghostkey` CLI / WASM / API built on top of it) uses "delegate" to mean the **rotated subordinate signing key** in the Ghost Key PKI trust chain:

```
Freenet master key ──signs──▶ delegate cert/key ──signs──▶ Ghost Key certificate
```

This naming predates Freenet's own [`Delegate`](https://freenet.org/build/manual/components/delegates/) concept — the sandboxed WASM agents that run inside a Freenet node, e.g. the `freenet/ghostkeys` wallet delegate. Those are the *correct* use of "delegate" and are what a technical reader reaches for when they hear the word today.

The collision is now actively confusing:

- In the upcoming `/ghostkey/` landing-page rewrite, the same paragraph ends up saying both "signed by the delegate key" (the PKI intermediate) and "held inside the Ghost Keys delegate" (the Freenet WASM agent). These are *completely different things*.
- `GhostkeyCertificateV1` has a field literally called `delegate: DelegateCertificateV1`, which to a `freenet-stdlib` reader sounds like it should be a `DelegateKey` pointing at a WASM component — it isn't.
- Search / grep over the Freenet source tree returns a mix of the two meanings.

## Proposed rename: `delegate` → `notary`

The ghost-key subordinate key's actual role: **it witnesses that a donation happened and signs an attestation to that effect**. Classic notary semantics — a trusted party witnesses a fact and publicly attests to it. Rationale for "notary" over alternatives (`issuer`, `intermediate`, `attestation`):

- **`intermediate`** (standard PKI term) is precise but jargon-y; doesn't read well in landing-page copy or error messages.
- **`issuer`** is semantically underspecified: the master is also an issuer (of the intermediate cert).
- **`attestation`** collides with TPM/remote-attestation terminology.
- **`notary`** has zero collision with anything in the Freenet stack, reads naturally to non-PKI readers, and accurately describes the role.

This issue tracks renaming the ghost-key subordinate-key concept everywhere inside the `freenet/web` repo (the `rust/` crates + CLI + docs). **It does not touch `freenet-core` or `freenet-stdlib`**, whose `Delegate*` types refer to the sandboxed WASM agent and are correctly named.

## ⚠️ Compatibility constraints — read before planning

Ghost Key certificates produced by the current code are **already out in the wild** (users have downloaded them). Any rename must be planned such that existing certs continue to verify with the renamed code, and ideally such that new code and old code can still exchange certs in both directions.

Two specific traps, both in `rust/gklib/src/armorable.rs`:

### Trap 1 — CBOR wire format is field-name-aware

`Armorable::to_bytes` / `from_bytes` serialise via **ciborium (CBOR)**, not bincode. CBOR encodes struct fields by name, so renaming a serde field in Rust **will break deserialisation** of existing files unless `#[serde(rename = "old_name")]` or `#[serde(alias = "old_name")]` is applied.

Concretely, today's wire format includes:

- `DelegateCertificateV1 { payload, signature }` — top-level fields `payload` and `signature` are fine; these don't contain the word "delegate".
- `DelegatePayload { delegate_verifying_key, info }` — the **`delegate_verifying_key` field name is in the wire format**. Renaming it in Rust without a serde rename will break every existing cert.
- `GhostkeyCertificateV1 { ..., delegate: DelegateCertificateV1, ... }` — the **`delegate` field name is in the wire format**. Same problem.

The agent picking this up must verify which field names are actually emitted by ciborium on the current code (a quick round-trip test is sufficient) and plan serde attributes accordingly.

### Trap 2 — PEM armor header/footer derives from Rust type name

`Armorable::struct_name()` uses `std::any::type_name::<Self>()` to generate the PEM armor identifier:

```rust
let full_name = type_name::<Self>();
// ...camel_case_to_upper...
// Produces e.g. "DELEGATE_CERTIFICATE_V1"
```

A `.pem` file written today has a header/footer like `-----BEGIN DELEGATE_CERTIFICATE_V1-----`. If `DelegateCertificateV1` is renamed to `NotaryCertificateV1`, the generated armor string becomes `NOTARY_CERTIFICATE_V1` and **`from_armored()` on an old file will fail the header check**.

This can't be fixed with `#[serde(rename)]` alone — it's a Rust-level type name lookup. The agent needs to either:

- Override `struct_name()` on the renamed type to return the old `DELEGATE_CERTIFICATE_V1` string (preserves backward compat but permanently ties the armor label to the old name — ugly but safe), OR
- Change `from_armored()` to accept a set of aliases `{"NOTARY_CERTIFICATE_V1", "DELEGATE_CERTIFICATE_V1"}` and strip/accept either on read, still writing with the new name.

The second option is cleaner long-term and should be the default choice unless there's a reason not to.

### Rust code (`rust/`)

| File | Usages |
|---|---|
| `rust/gklib/src/delegate_certificate.rs` | 27 |
| `rust/gklib/src/ghost_key_certificate.rs` | 38 |
| `rust/gkwasm/src/lib.rs` | 34 |
| `rust/cli/src/bin/ghostkey.rs` | 31 |
| `rust/cli/src/commands.rs` | 35 |
| `rust/api/src/delegates.rs` | 9 |
| `rust/gklib/src/lib.rs`, error variants, tests | — |

Concrete renames (non-exhaustive — `git grep -n -w -i delegate rust/` gives the full list):

- Types: `DelegateCertificateV1` → `NotaryCertificateV1`, `DelegatePayload` → `NotaryPayload`
- Fields: `delegate_verifying_key` → `notary_verifying_key` (**with `#[serde(alias = "delegate_verifying_key")]`**), `delegate: DelegateCertificateV1` on `GhostkeyCertificateV1` → `notary: NotaryCertificateV1` (**with `#[serde(rename = "delegate")]` or alias**)
- Function/variable names: `delegate_signing_key`, `delegate_keypair`, `delegate_certificate`, etc.
- Error variants referencing "delegate"
- Module: `rust/gklib/src/delegate_certificate.rs` → `rust/gklib/src/notary_certificate.rs` (with a re-export from the old path for one release cycle if easy)

### CLI subcommands (`rust/cli/`)

Current public CLI surface uses `delegate` directly:

```
ghostkey generate-delegate --master-signing-key ... --info ...
ghostkey generate-ghost-key --delegate-cert ... --delegate-key ...
```

Plan:

- Add new subcommands: `generate-notary`, `--notary-cert`, `--notary-key`, etc.
- **Keep the old `delegate` subcommand/flags as deprecated aliases** for at least one minor release. Emit a deprecation warning to stderr on first use.
- Update `test_cli_commands.sh` to exercise both the new and the legacy names.

### API (`rust/api/`)

- `rust/api/src/delegates.rs` — rename the module, any Axum route paths, any JSON field names in request/response bodies. API wire format is a second compatibility surface: if any endpoint is publicly called (e.g. from the donation flow in `gkwasm`), plan aliases there too.

### WASM (`rust/gkwasm/`)

- `wasm_bindgen`-exposed function names currently include `delegate`. These are called from `hugo-site/content/ghostkey/create/index.html`'s JavaScript — coordinate the rename or keep aliases so the old JS keeps working until the HTML is updated in the same PR.

### Hugo content (`hugo-site/`)

- `hugo-site/content/ghostkey/_index.md` — new landing-page copy (currently on branch `ghostkeys-rewrite`) uses `delegate key` for the PKI subordinate; rename to `notary certificate` / `notary key`.
- `hugo-site/layouts/shortcodes/ghostkeys-diagram-blindsign.html` — diagram says "Freenet server". Codex review recommended clarifying; "Freenet notary" or "Donation notary" is a good replacement label.
- `hugo-site/layouts/shortcodes/ghost-key-explanation.html` (legacy) — check if still referenced; if yes, rename; if not, delete in this PR.
- `hugo-site/content/ghostkey/create/` and `hugo-site/content/ghostkey/success/` — check for `delegate`-flavoured user-facing copy.
- `rust/gklib/AGENTS.md` — update the "Core Flow" section (step 3) and "Useful CLI Commands" section.

### Docs

- `rust/gklib/README.md` — currently describes `DelegateCertificateV1`. Update.
- Crate-level rustdoc on `ghostkey_lib`.
- Any external docs under `hugo-site/content/` that link to the delegate concept.

### Explicitly out of scope (do not touch)

- Anything in `freenet-core`, `freenet-stdlib`, `freenet/ghostkeys` (the wallet delegate), or other Freenet repos. Their `Delegate*` types are the correct, current, landed meaning of "delegate".
- The old `/about/news/introducing-ghost-keys/` blog article can be left alone as historical; consider adding a one-line editor's note at the top if the terminology shift is confusing on reread.

## Migration strategy

Recommended approach: **source rename + wire-format aliases, no format bump**.

1. **Rename Rust identifiers** across all files listed above.
2. **Preserve wire format** via `#[serde(rename = "...")]` / `#[serde(alias = "...")]` so existing CBOR bytes round-trip unchanged.
3. **Preserve PEM armor** by overriding `Armorable::struct_name()` for the renamed types, OR teach `from_armored()` to accept a set of aliases. Prefer the latter.
4. **Publish as `ghostkey_lib = 0.1.5`** (patch bump — this is a source-API rename and we ship type aliases for the old names, so downstream code that names `DelegateCertificateV1` still compiles with a deprecation warning).

Explicitly rejected alternatives:

- **Introduce `NotaryCertificateV2` as a new wire format** — unnecessary churn. The existing format is fine; only the naming was wrong.
- **Full break without aliases** — burns existing users and certs for a purely cosmetic win. No.

## Deprecation shims

Add to `rust/gklib/src/lib.rs`:

```rust
#[deprecated(since = "0.1.5", note = "renamed to NotaryCertificateV1")]
pub type DelegateCertificateV1 = NotaryCertificateV1;

#[deprecated(since = "0.1.5", note = "renamed to NotaryPayload")]
pub type DelegatePayload = NotaryPayload;
```

Plus equivalent function-level `#[deprecated]` re-exports for any public helpers that embed "delegate" in their name.

Plan to remove deprecated aliases in `0.2.0` or later.

## Test plan

This is the critical section. The rename must be verified against real existing certificates.

1. **Before starting**: on `main`, produce a set of test certificates at HEAD and save them to `rust/gklib/tests/fixtures/legacy_v1/`:
   - A master-key-signed delegate cert
   - A ghost key certificate
   - A signed message
2. **After renaming**: add a test `test_legacy_v1_fixtures_still_verify` in `rust/gklib/tests/` that loads each fixture and asserts:
   - `NotaryCertificateV1::from_file()` succeeds on the legacy `.pem`
   - `.verify()` succeeds against the master key
   - `GhostkeyCertificateV1::from_file()` succeeds on the legacy ghost cert
   - `.verify()` chains back to the master correctly
   - A signed message produced by old code verifies against the new code
3. **Round-trip test**: load a legacy cert, re-serialise it with new code, reload it — bytes should be identical (if serde rename is applied correctly) or at minimum the re-loaded struct should equal the original.
4. **CLI compat**: `test_cli_commands.sh` must exercise both `generate-delegate` (deprecated alias) and `generate-notary`, and the output of each must be readable by `verify-ghost-key` under both old and new flag names.
5. **Integration test**: the existing `cargo make integration-test` pipeline (end-to-end Stripe → sign → verify flow) must pass unchanged.
6. **WASM smoke test**: rebuild `gkwasm` and manually run `hugo-site/content/ghostkey/create/` locally — the donation flow must still produce a valid certificate.

## Urgency and scope

**Do this in a single PR and land it soon.** Ghost Keys are still experimental and very few certificates are in the wild today, so the compat surface is minimal right now — but it grows rapidly as adoption picks up. The later this rename ships, the more real users it will affect. One clean PR now is cheaper than a multi-phase migration later.

The two compat traps (CBOR field names, PEM armor string) are both cheap to handle — `#[serde(alias)]` attributes and a small tweak to `from_armored` — so do them as part of the same PR. They are not a reason to split the work.

## Acceptance criteria

- [ ] All Rust identifiers in `rust/gklib`, `rust/gkwasm`, `rust/cli`, `rust/api` that refer to the PKI subordinate concept use `Notary` / `notary` instead of `Delegate` / `delegate`.
- [ ] Deprecated type aliases exist for `DelegateCertificateV1`, `DelegatePayload`.
- [ ] Fixture tests prove that `.pem` certs produced by the pre-rename code still verify with post-rename code.
- [ ] CLI supports both `generate-notary` (new) and `generate-delegate` (deprecated) with equivalent behaviour.
- [ ] `cargo test --all`, `cargo clippy --all-targets`, `cargo fmt --check`, `./test_cli_commands.sh`, and `cargo make integration-test` all pass.
- [ ] Hugo site copy under `/ghostkey/` uses "notary" terminology consistently; the new landing page on branch `ghostkeys-rewrite` is updated before merge.
- [ ] `rust/gklib/README.md`, `rust/gklib/AGENTS.md`, and rustdoc describe the notary chain accurately.
- [ ] **No changes** to `freenet-core`, `freenet-stdlib`, or `freenet/ghostkeys` — their `Delegate*` types remain untouched.

## Estimated surface area

~174 occurrences of the string "delegate" across six Rust files, plus CLI flags, JSON field names in the API, the PEM armor string, and Hugo content. Most changes are mechanical; the risk concentrates in the two compatibility traps above.

[AI-assisted - Claude]

sanity:
## Problem

`ghostkey_lib` uses "delegate certificate" for the PKI intermediate signing
key in the ghost-key trust chain (`master → delegate → ghost`). This collides
with Freenet's own [`Delegate`](https://freenet.org/build/manual/components/delegates/)
concept — the sandboxed WASM agents like `freenet/ghostkeys`. The collision
makes docs, code, and conversation ambiguous. The new `/ghostkey/` landing
page copy even ends up saying both "signed by the delegate key" (PKI) and
"held inside the Ghost Keys delegate" (WASM agent) in the same paragraph.

Tracking issue: #24.

## Approach

Rename "delegate" → "notary" throughout `ghostkey_lib`, the `ghostkey` CLI,
the `ghostkey-api` server, `gkwasm`, and the Hugo site. **Source API only** —
ghost-key certificates minted by 0.1.4 and earlier continue to parse, verify,
and round-trip byte-identically under 0.1.5.

Alternatives considered and rejected:

- **Bump to a V2 cert format with new field names**. Rejected as unnecessary
  churn (as called out in the issue itself). Cert signatures cover the
  serialized payload bytes, so any CBOR field rename would invalidate every
  existing ghost key. We freeze the wire format via `#[serde(rename)]` and
  only rename the Rust source identifiers.
- **Full break without aliases**. Rejected — burns existing users and certs
  for a purely cosmetic win.

### Compatibility shims (all planned for removal in 0.2.0)

| Shim | Location |
|---|---|
| `DelegateCertificateV1` / `DelegatePayload` deprecated type aliases | `rust/gklib/src/delegate_certificate.rs` stub module |
| `#[serde(rename = "delegate_verifying_key")]` freezing CBOR field | `NotaryPayload` |
| `#[serde(rename = "delegate")]` freezing CBOR field | `GhostkeyCertificateV1` |
| PEM armor legacy header acceptance (`BEGIN DELEGATE_CERTIFICATE_V1`) | `rust/gklib/src/armorable.rs` `legacy_armor_aliases()` |
| Clap aliases: `generate-delegate`, `verify-delegate`, `--delegate-certificate`, `--delegate-dir` | `rust/cli/src/bin/ghostkey.rs`, `rust/api/src/main.rs` |
| Deprecation warnings via `argv` pre-scan (clap normalizes matches) | `rust/cli/src/bin/ghostkey.rs::warn_on_legacy_argv`, `rust/api/src/main.rs` |
| Legacy on-disk filename fallback (`delegate_certificate.pem` → `notary_certificate.pem`) | `rust/cli/src/commands.rs::resolve_notary_file` |
| Legacy per-amount filename fallback (`delegate_{kind}_{amount}.pem`) | `rust/api/src/delegates.rs::resolve_amount_file` |
| `DELEGATE_DIR` env var fallback | `rust/api/src/delegates.rs::notary_dir` |
| `generate_delegate_keys.sh` stub execing `generate_notary_keys.sh` | `rust/cli/` |
| Dual-emit `delegate_certificate_base64` + `notary_certificate_base64` JSON fields on `DonationResponse` and `SignCertificateResponse` | `rust/api/src/routes.rs`, `rust/api/src/handle_sign_cert.rs` |
| localStorage read fallback + migration (don't delete old key) | `hugo-site/static/js/donation-success.js` |
| localStorage dual-write | `hugo-site/themes/freenet/layouts/shortcodes/stripe-donation-form.html` |

### Deliberately NOT touched

- `hugo-site/content/about/news/introducing-ghost-keys.md` — historical blog
  post. The embedded JSON key `delegate-key-created` is baked into real cert
  `info` fields of every donation ever made.
- `generate_notary_keys.sh` still emits `"delegate-key-created"` as a JSON
  key inside cert info. Parsed by River's UI and by wild historical certs.
- `rust/api/src/delegates.rs` module filename — internals renamed, doc
  comment explains. Moving the file would churn the module path for no win.

### Sequencing

This PR lands in `freenet/web`. After it merges and `ghostkey_lib 0.1.5`
publishes to crates.io, a follow-up PR will land in `freenet/ghostkeys` to
rename `common/src/lib.rs`, `ghostkey-delegate` handlers, integration tests,
and UI consumers. The ghostkeys PR is out of scope here — this PR's
compatibility shims ensure ghostkeys keeps compiling against 0.1.5 unchanged.

## Testing

**Fixture-based regression tests** (`rust/gklib/tests/legacy_v1_compat.rs`)
are the line of defense. Fixtures were generated from an unmodified checkout
of `main` in a scratch worktree *before* any rename commits, then committed
alongside the tests:

- `tests/fixtures/legacy_v1/master_verifying_key.pem`
- `tests/fixtures/legacy_v1/delegate_certificate.pem` (with `BEGIN DELEGATE_CERTIFICATE_V1` header)
- `tests/fixtures/legacy_v1/ghost_key_certificate.pem`
- `tests/fixtures/legacy_v1/signed_message.bin`

The tests assert:

1. Legacy PEM parses under the new types via armor alias
2. Legacy ghost key cert verifies back to the fixture master key
3. Legacy `SignedMessage` still verifies (CBOR deserialize + ed25519 sig)
4. Re-serializing a loaded legacy cert is **byte-identical** to the fixture
   (strongest possible check that the `#[serde(rename)]` freeze is correct —
   any drift would change payload bytes and break the master signature)
5. New code writes canonical `BEGIN NOTARY_CERTIFICATE_V1` headers

**CLI smoke tests** exercise both spellings:

- `test_cli_commands.sh`: 17/17 pass (includes legacy `generate-delegate` /
  `verify-delegate` / `--delegate-certificate` alias tests). Also fixed a
  pre-existing `set -e` + `((TOTAL_TESTS++))` footgun that was causing the
  script to silently exit before running any tests.
- `rust/cli/test_ghostkey.sh`: 19/19 pass (includes legacy `--delegate-dir`
  alias tests)

**Per-crate test/check runs:**

- `cargo test --manifest-path rust/gklib/Cargo.toml --all-features` — 22 unit
  + 6 regression fixture tests green
- `cargo test --manifest-path rust/Cargo.toml --all-features` — api +
  integration_test + gkwasm green
- `cargo test --manifest-path rust/cli/Cargo.toml --all-features` — green

**Not run locally:**

- `cargo make integration-test` (full Stripe → sign → verify e2e) — requires
  Stripe secret keys and a running Hugo dev server. This is the highest-risk
  gap in local validation, and reviewers should treat CI's run of it as the
  authoritative signal.
- Manual browser smoke test of the donation flow — recommend reviewer
  spot-check this after merge.

## Release checklist note

The `ghostkey-api` and `integration_test` crates previously depended on
`ghostkey_lib` from crates.io and now depend on the in-tree `../gklib` path.
This removes an implicit check that the published crate doesn't drift from
the in-tree source. After publishing 0.1.5, manually verify by temporarily
flipping `rust/api/Cargo.toml` to `ghostkey_lib = "0.1.5"` and running
`cargo check --manifest-path rust/api/Cargo.toml`. See `rust/gklib/CHANGELOG.md`.

## Commit structure

Nine logical commits, each independently reviewable:

1. `refactor(workspace)` — path-dep switch so rename commits can compile
2. `refactor(gklib)` — core rename + serde freeze + armor aliases
3. `test(gklib)` — legacy fixture regression tests (ran IMMEDIATELY after
   the rename to validate the wire-format decisions before touching anything
   else)
4. `refactor(gkwasm)` — internal rename
5. `refactor(cli)` — subcommands + flags + filenames + fallbacks
6. `refactor(api+scripts)` — api + shell scripts + env vars + dual-emit JSON
7. `docs(web)` — Hugo + JS + READMEs + AGENTS.md
8. `chore(gklib)` — version bump + CHANGELOG
9. `style` — `cargo fmt` on touched files

Closes #24

🤖 [AI-assisted - Claude]


sanity:
Addressed review round 1 findings in commit cc3b30f8:

**Version bump 0.1.5 → 0.2.0** (codex P2 / breaking-semver correctness). Renaming `GhostkeyCertificateV1.delegate` → `.notary` is source-breaking for downstream field accessors. Under pre-1.0 Rust semver, `0.1.x → 0.1.y` is the compatible range — Cargo would auto-upgrade downstream crates and break their builds without opt-in. `0.2.0` is the breaking range and requires explicit dep bump. CHANGELOG, README, AGENTS.md, all deprecation `since` attrs, and all "removed in 0.2.0" forward-references rewritten to "a future release".

**Skeptical HIGH #1** — `--notary-dir` required-without-env bug. Operators running with only legacy `DELEGATE_DIR` set (systemd units / `.env` files) would hit a hard clap error because the `delegates::notary_dir` fallback was unreachable. Fixed: pre-parse hook in `main()` hydrates `NOTARY_DIR` from `DELEGATE_DIR` before clap runs, with a deprecation warning. `--notary-dir` is now also bound to `.env("NOTARY_DIR")`.

**Skeptical MEDIUM #3** — `resolve_amount_file` pair-atomicity bug. Resolved cert and signing-key independently → partial on-disk migration could pair a new-named cert with a legacy-named signing key, producing ghost certs with broken signature chains. Replaced with `pick_scheme()` which resolves the naming scheme at the directory level. Six unit tests pin every branch including both partial-migration cases.

**Codex LOW #7** — `rust/gklib/Cargo.toml` `repository` URL was pre-existing wrong (`freenet-core`). Fixed to `freenet/web`.

**Testing reviewer HIGH #1** — no automated tests for the fallback resolvers. Added 4 tests for `resolve_notary_file` in `rust/cli/src/commands.rs` and 6 tests for `pick_scheme` + `NamingScheme` in `rust/api/src/delegates.rs`.

**Terminology** — updated AGENTS.md and CHANGELOG to call out three distinct concepts: **notary certificate** (PKI intermediate, this crate), **Freenet `Delegate`** (generic platform primitive), and **Ghostkey Vault** (the specific product name of the `freenet/ghostkeys` delegate — issue #24's "wallet delegate" wording is superseded; use "vault" going forward).

**Minor cleanups:**
- `browser_test.rs:220` now accepts both legacy and new error message strings.
- `environment.rs` dropped dead `GHOSTKEY_DELEGATE_KEY_DIR` env var write (nothing reads it).
- `#[deprecated(since = "0.1.5", ...)]` → `since = "0.2.0"`.
- Inline "0.1.5" doc comments synced to 0.2.0.

Test status after the fixes (all green locally):
- `cargo test -p ghostkey_lib` — 22 unit + 6 legacy fixture tests
- `cargo test -p ghostkey-api` — 14 tests (includes 6 new `pick_scheme` tests)
- `cargo test -p gkwasm` — 3/3
- `cargo test -p ghostkey` (cli) — 4 new `resolve_notary_file` tests
- `test_cli_commands.sh` — 17/17 (both new and legacy spellings)
- `test_ghostkey.sh` — 19/19

[AI-assisted - Claude]

PR #35, opened 2026-04-18
docs: add uninstall guidance to quickstart and FAQ

sanity:
## Problem

A user on Matrix tried to uninstall Freenet with `sudo freenet uninstall` and got no output. Cause: the `curl -fsSL https://freenet.org/install.sh | sh` installer drops the binary in `~/.local/bin`, which isn't on `sudo`'s `secure_path`, so the command silently fails. `whereis` and `locate` don't cover user-local binaries either, so the user concluded Freenet was unremovable and left in frustration.

Our site currently has no user-facing documentation on how to uninstall — the install page tells people how to put Freenet on their machine but nothing about how to take it off.

## Approach

Defense-in-depth, documentation layer. Two additions:

1. **Quick Start guide** — an "Uninstalling" section explaining:
   - The `freenet uninstall` command and its `--purge` / `--keep-data` flags.
   - Why you should **not** prefix it with `sudo` for a normal install.
   - How to call it by full path if `~/.local/bin` isn't on `$PATH`.
   - How to clean up a `cargo install freenet` install.
   - A manual-fallback shell snippet that deletes the binaries, the systemd user unit, and the XDG data/config/cache/log directories when the binary is missing or broken. macOS paths noted.

2. **FAQ entry** at `/about/faq/#how-do-i-uninstall-freenet` covering the same ground more concisely and linking back to the Quick Start for the full walkthrough. Also added to the FAQ's in-page TOC.

Companion changes to freenet-core (standalone `uninstall.sh` script, clearer post-install message, having `freenet uninstall` also find `cargo install`-placed binaries) will follow in a separate PR.

## Testing

- `hugo` builds the site cleanly with the new content.
- Verified rendered HTML contains `<h2 id="uninstalling">` on the Quick Start page and `<h1 id="how-do-i-uninstall-freenet">` in the FAQ, so the new anchors are reachable and the Quick Start cross-link from the FAQ resolves.
- `git diff --stat` is 62 lines, all additions — no reformatting of unrelated content.

## Fixes

None — user-reported on Matrix, not tracked as a GitHub issue.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/content/about/faq/index.md
+++ b/hugo-site/content/about/faq/index.md
@@ -20,6 +20,7 @@ aliases:
 - [What is the status of Freenet?](#what-is-the-status-of-freenet)
 - [Can I follow Freenet on social media?](#can-i-follow-freenet-on-social-media)
 - [How can I financially support Freenet development?](#how-can-i-financially-support-freenet-development)
+- [How do I uninstall Freenet?](#how-do-i-uninstall-freenet)
 - [Why does the Freenet project use and mention AI tools?](#why-does-the-freenet-project-use-and-mention-ai-tools)
 
 Freenet is a fully decentralized, peer-to-peer network and a drop-in replacement for the world wide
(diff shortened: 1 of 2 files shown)


sanity:
## Review responses (code-first + big-picture + Codex + Matrix feedback from @Ivvvor)

Addressed findings in 2 follow-up commits (d3ce5f5 for Windows, 852002a for macOS fixes):

### Code-first reviewer findings

**macOS paths were wrong (critical):** the `directories` v6 crate on macOS uses a bundle-ID layout (`The-Freenet-Project-Inc.Freenet`), not a bare `Freenet` directory. Every macOS path in the original manual fallback pointed at folders that never existed. Rewrote the macOS section with the correct paths (both uppercase `.Freenet` and lowercase `.freenet` cache variants) and added `~/Library/Logs/freenet` (lowercase) for logs.

**`--keep-data` wording:** tightened — it skips the entire purge step, so config/cache/logs are preserved alongside data. Updated both pages.

**Sudo failure mode:** softened to the specific `command not found` symptom matching the Matrix report, and noted the system-wide unit file path at `/etc/systemd/system/freenet.service` for users who chose `--system`.

**`cargo uninstall fdev` conditional:** kept — `cargo install freenet` doesn't pull `fdev`, so the conditional is correct.

### Big-picture reviewer findings

**macOS launchd teardown missing:** added `launchctl unload` and plist removal lines — the service is a launchd user agent, not systemd, so the original snippet would have leaked the agent.

**Windows not mentioned:** added a full Windows section (also prompted by @Ivvvor's Matrix report — see below).

**FAQ ordering / `os-install` companion shortcode / post-install message:** deferred to follow-up. FAQ ordering is a nice-to-have and the post-install message update belongs in the companion freenet-core PR the description already references.

### Matrix feedback (@Ivvvor)

Reported that `freenet uninstall` on Windows leaves the config folder behind (and the installer-placed `%LOCALAPPDATA%\Freenet\bin\` is a different path from what `uninstall` detects). Added:

- PowerShell snippet covering `%LOCALAPPDATA%\Freenet\bin`, the data/logs tree under `%LOCALAPPDATA%`, and the config tree under `%APPDATA%`.
- Note about checking `HKCU:\Software\Microsoft\Windows\CurrentVersion\Run` for leftover autostart entries.
- Short Windows pointer in the FAQ linking back to the full snippet.

Filed underlying code bug as freenet/freenet-core#3904 for the companion core PR to address.

### Codex

Found no discrete correctness issues.

### Other deferred items

- Standalone `uninstall.sh` script, post-install message update, and cargo-bin-aware detection all land in the companion freenet-core PR (tracked separately, not in this PR's scope).

[AI-assisted - Claude]

Issue #5, opened 2024-10-10
improove modularity of frenet site theme


gogo2464:
I would like to reuse https://github.com/freenet/web/tree/main/hugo-site/themes/freenet .

but I can not do ` hugo mod init https://github.com/freenet/web/tree/main/hugo-site/themes/freenet` in a single line. Could we split the repo with a clonable theme with command ` hugo mod init https://github.com/freenet/web/` please?

sanity:
# Issue Priority Assessment

Generated: 2026-02-18T22:11:42.558417590+00:00
Issue hash: `ef21952b`

## Value Assessment

There is a modest but real user impact here for people trying to reuse the Freenet Hugo theme as a standalone module: the theme lives inside `hugo-site/themes/freenet` alongside the full site, so the `hugo mod init https://github.com/freenet/web/tree/main/hugo-site/themes/freenet` command fails because Hugo modules expect a repository root rather than a nested subdirectory. Anyone who wants to adopt just the look-and-feel for a fork, mirror, or related site therefore has to clone the entire `freenet/web` tree and manually extract the theme, which adds friction and invites duplication. The severity is low–it doesn’t break the existing site–but it does make theme reuse harder for community members or downstream projects whose primary goal is to share the theme rather than the whole website.

Strategically, the Freenet website’s objectives center on maintainability and the reuse of tooling (the layout and Ghost Key integration being common elements). Making the theme easily importable as a Hugo module would reinforce that strategy by letting other repositories import it via `module` semantics instead of copying files. Because the Hugo site already references its theme through `hugo-site/config.toml` and keeps the theme content in one place, splitting the theme into its own repository or at least making the repo root align with Hugo module expectations would keep the main site focused on content while supporting downstream contributors who want to build on the same visual system.

This request isn’t blocking any critical path work—core functionality like Ghost Key WASM builds (`rust/gkwasm`), the Axum API, or Hugo content updates can continue while the theme remains nested. No other open major contributions appear to be waiting on this modularization, so it doesn’t hold up releases or infrastructure changes. That said, future contributions that rely on the theme as a dependency would benefit, so clearing this hurdle sooner keeps those workflows unblocked.

There’s no urgent timing pressure mentioned, and nothing in the codebase indicates a security or stability risk tied to the current setup. Addressing it is nice-to-have structural cleanup; keeping it in the backlog with a “help wanted” tag is likely sufficient unless someone actively needs to consume the theme as a Hugo module.

## Cost Assessment

[...]

PR #42, opened 2026-04-30
fix(install): read prompt from /dev/tty when stdin is piped

sanity:
## Problem

The Linux install script advertised on the quickstart page is invoked via
`curl -fsSL https://freenet.org/install.sh | sh`. When piped, stdin is the
curl output (script bytes), not the user's terminal, so `read -r response`
at the service-install prompt silently consumes nothing and the prompt
ignores keyboard input.

Reported by Ivvor on Matrix:
> @Ian: Freenet linux install script is broken. Streaming curl to sh causes
> it to not process keyboard input when asking if you want the service
> installed.

## Approach

A previous fix (`eff317ae`, Dec 2025) wrote `read -r response < /dev/tty`,
but PRs #36 / #37 ("resync install.sh + uninstall.sh with freenet-core")
overwrote the file with the freenet-core copy and dropped the override.

This change reapplies the fix and hardens it:

- **stdin is a TTY** → `read` directly (covers `sh install.sh`).
- **stdin is piped, /dev/tty openable** → `read < /dev/tty` (covers `curl | sh`).
- **no tty at all** → skip the prompt with an info message rather than
  aborting under `set -eu` (covers CI / `setsid`).

`{ true </dev/tty; } 2>/dev/null` is used instead of `[ -r /dev/tty ]`,
because the `-r` access check can succeed when the process has no
controlling tty and the subsequent open fails with ENXIO.

## Testing

Smoke-tested with a small harness that extracts the prompt block:

- **Case A — `curl | sh` under a real pty (`script -c`)**: stdin is piped,
  /dev/tty present. Script reads `y` from /dev/tty and proceeds with
  service install. ✅
- **Case B — `setsid ... </dev/null` (no controlling tty)**: script prints
  "Non-interactive shell detected; skipping service installation prompt"
  and continues without erroring. ✅
- `sh -n` parses cleanly.

## Follow-up

`freenet-core/scripts/install.sh` carries the identical bug (the two files
are bit-identical). A matching PR will land there next so the next resync
doesn't regress this file again.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/static/install.sh
+++ b/hugo-site/static/install.sh
@@ -409,11 +409,26 @@ main() {
         print_path_instructions "$install_dir"
     fi
 
-    # Ask about service installation (unless FREENET_NO_SERVICE is set)
+    # Ask about service installation (unless FREENET_NO_SERVICE is set).
+    # When the script is piped via `curl ... | sh`, stdin is the pipe carrying
+    # script bytes, not the terminal, so a plain `read` cannot capture
+    # keystrokes. Read from /dev/tty when stdin isn't a TTY; if no TTY is
+    # available at all (truly non-interactive shell), skip the prompt rather
+    # than failing under `set -eu`. Note: `[ -r /dev/tty ]` is unreliable —
+    # the access check can pass even when the process has no controlling tty
+    # and the subsequent open fails. Probe by actually opening /dev/tty.
     if [ "${FREENET_NO_SERVICE:-0}" != "1" ]; then
         echo ""
-        printf "Would you like to install Freenet as a system service? [y/N] "
-        read -r response
+        response=""
+        if [ -t 0 ]; then
+            printf "Would you like to install Freenet as a system service? [y/N] "
+            read -r response
+        elif { true </dev/tty; } 2>/dev/null; then
+            printf "Would you like to install Freenet as a system service? [y/N] "
+            read -r response </dev/tty
+        else
+            info "Non-interactive shell detected; skipping service installation prompt."
+        fi
         case "$response" in
             [yY]|[yY][eE][sS])
                 info "Installing service..."
--- a/hugo-site/static/install.sh
+++ b/hugo-site/static/install.sh
@@ -9,6 +9,14 @@
 #   FREENET_INSTALL_DIR  - Installation directory (default: ~/.local/bin)
 #   FREENET_NO_SERVICE   - Set to 1 to skip service installation prompt
 #   FREENET_VERSION      - Specific version to install (default: latest)
+#
+# NOTE: A near-identical copy lives at freenet-core/scripts/install.sh.
+# Keep the two in sync. In particular, the service-install prompt
+# (look for "Would you like to install Freenet as a system service")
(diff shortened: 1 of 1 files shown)


sanity:
## Review-feedback response

Ran 4 internal reviewers (code-first, testing, skeptical, big-picture) and Codex CLI against the initial commit. Summary of how each finding was handled in the follow-up commit `de6e42c7`:

### Addressed

| # | Source | Finding | Resolution |
|---|--------|---------|------------|
| 1 | Codex P2 | `[ -t 0 ]`-only check makes `printf 'y' \| sh install.sh` automation block on /dev/tty. | Switched to `case "${0##*/}"` discriminator: if `$0` is a shell name (`sh`/`bash`/etc.), the script source came from stdin (`curl \| sh`), so read from /dev/tty; otherwise stdin carries the answer, read from stdin. Verified `printf 'y' \| sh install.sh` still works. |
| 2 | Skeptical #1 | `read` EOF (Ctrl-D, closed stdin) aborts under `set -eu`. | Both `read` calls now `\|\| response=""`. |
| 3 | Skeptical #10 | Em-dash in comment violates writing-style rule. | Removed; comment block rewritten without em-dashes. |
| 4 | Skeptical #2/#3 | Silent skip in non-interactive contexts gives no remediation hint. | `*)` case arm now appends "To skip this prompt entirely in scripted installs, set FREENET_NO_SERVICE=1." |
| 5 | Skeptical #7 | Redundant "Non-interactive shell detected" + "Skipping service installation" messages. | Removed the info line; the `*)` case arm already prints the skip message. |
| 6 | Big-picture #5 | No deterrent to a future resync silently dropping the fix again. | Added a NOTE block at the top of `install.sh` cross-referencing `freenet-core/scripts/install.sh` and naming this PR + the prior regression. |

### Deferred (worth doing, separate PR scope)

| # | Source | Finding | Why deferred |
|---|--------|---------|------------|
| - | Testing #1 | CI grep-guard against the regressed two-line form. | Highest-leverage testing item, but it's CI infrastructure — better as its own PR alongside the freenet-core fix. |
| - | Testing #2/#3 | Extract prompt block into a function + unit test, plus a `setsid sh -c 'cat install.sh \| sh'` end-to-end smoke test. | Refactor + new test infrastructure; out of scope for a hotfix. |
| - | Skeptical #4 | Behavior on busybox ash (Alpine). | Would need a CI matrix that doesn't exist; relying on POSIX-defined `{ ...; } 2>/dev/null` semantics for now. If anyone reports issues on Alpine we fix them then. |

### Manual verification

Re-ran the pty harness over five scenarios (all pass):
1. `curl | sh` under pty: reads `y` from /dev/tty, installs service.
2. `sh install.sh` interactive (heredoc): reads `y` from stdin.
3. `printf 'y' | sh install.sh` (Codex's case): reads `y` from piped stdin.
4. `setsid sh -c './install.sh' </dev/null`: skips prompt, prints `FREENET_NO_SERVICE=1` hint.
5. `sh install.sh </dev/null`: handles EOF without aborting under `set -eu`.

Companion PR against `freenet-core/scripts/install.sh` is next.

[AI-assisted - Claude]

PR #87, opened 2026-07-27
feat: lead the homepage with River instead of burying it

sanity:
## Problem

River is the strongest thing the project ships, but finding it from the landing page meant digging through the Apps dropdown. River was never named above the fold, and the homepage carried no product imagery at all, so a visitor could read the entire page without ever seeing what a Freenet app looks like.

The one-click River path was **already in the hero**. It was the link labelled "Or try it in your browser first", which goes to `/try/`, which opens River and drops you into the live Freenet Official room with no install and no account. Nothing said so. It was styled as a bare text link beside a filled button, and "try it **first**" framed it as the cautious option rather than the shortest path to a working app.

Separately the page felt busy. Length was not the problem (1,615px, under two viewports). The crowding was one band: directly under the hero, How Freenet Works, News and Watch all competed at equal weight, and the routing diagram was the largest element on the page, answering question three before question one.

## Approach

**Hero.** Two columns: copy left, a real River screenshot framed in browser chrome right, linking to `/river/`. Reuses the existing light/dark screenshot assets already on the River page. The copy names River and says what it is ("group chat that feels like Signal or WhatsApp with nobody in the middle") rather than stacking four abstractions into one sentence.

**The River button, and a deliberate constraint.** The old CSS carried this comment:

> Secondary CTA: the hosted-demo "try in your browser" entry. Kept visually subordinate to the primary Install button so the real (self-hosted) path stays front and center.

That is a real and correct concern: `/try/` is a centralized hosted demo, and promoting it as *the* primary CTA on a decentralization project's homepage would undercut the message. So rather than flipping the hierarchy, the River entry becomes an outlined button of the same size and hit target as Install. Install stays the filled primary. A new note under the row says plainly that the browser version is a hosted demo and that installing a peer gives you the real thing. Findable, without misrepresenting what it is.

**Order.** Audience cards moved above How Freenet Works. Lead with what you can do, follow with how it works.

**Headline.** Moved into the hero grid so it shares a column with the copy and CTAs (`home.html` no longer emits it). Other layouts are untouched and still render `.Title` themselves; verified `/about/papers/` and `/quickstart/` still render their own `h1`. The `<title>` tag is unchanged.

**Two content bugs.** "For Users" said "the easiest place to start is River" and then pointed its CTA at `/quickstart/`; it now goes where the sentence promises. News trimmed from 5 to 3 — the `limit` param existed but the release-merging JS path ignored it and hardcoded `slice(0, 5)`.

## Fixes found on the way

- **Live-dashboard link** pointed at `http://nova.locut.us:3133/`, a raw host:port over plain HTTP that 301s to `https://telemetry.freenet.org/` anyway. Now links the canonical HTTPS URL directly.
- **320px horizontal scroll**, from two independent causes:
  - the nav burger was pushed off the right edge by the logo plus Donate button. **Pre-existing on `main`** (14px), fixed here since it is the same page.
  - the screenshot's `nowrap` address bar set a min-content width on its flex item, which propagated up through the grid column (13px, introduced by this PR). Fixed with `min-width: 0` plus explicit shrink, with a comment saying why it is load-bearing.
  - CTAs also go full-width below 360px so the longer label cannot force a sideways scroll.

## Testing

Rendered and measured at **320, 390, 430, 768, 1024, 1280, 1440 and 1920px** wide, in **both light and dark** (16 combinations):

- no horizontal scroll and no element overflowing the viewport at any size
- the River CTA is above the fold at every size
- News renders exactly 3 items everywhere
- section order confirmed as hero → cards → how-it-works/news → secondary links

Also verified the network ring still animates. It appears blank in a full-page screenshot because an `IntersectionObserver` gates the draw loop, which is pre-existing intended behaviour and simply more visible now that the section sits below the fold; it renders on scroll, confirmed by sampling non-transparent canvas pixels.

`hugo --gc --minify` builds clean, 122 pages. `content/_index.md` now passes `prettier --check` (it was already dirty on `main`).

[AI-assisted - Claude]


sanity:
## Review findings addressed (6e62a45)

An independent adversarial review found no blockers but eight should-fixes. All are fixed. Numbers below are re-measured, not quoted from the review.

### CLS regression, the strongest finding

Centering the hero grid turned the webfont swap into a vertical translation of the screenshot and everything below it. Measured **0.121** at 1280px against **0.020** on `main`, over Google's 0.1 threshold. The reviewer isolated the cause properly (blocking the PNG left CLS at 0.115; blocking the webfonts took it to 0), so this was the `display=swap` FOUT amplified by `align-items: center`.

Fixed by anchoring the grid to the top and giving the screenshot a fixed `26rem` height with `object-fit: cover`, so the copy column reflowing no longer moves anything below it. Re-measured under the same 1.6Mbps/150ms throttle: **0.063 light, 0.063 dark**.

### Image weight

The hero shipped a 202KB PNG to fill at most 552 CSS px. Added WebP variants sized for the hero, with the original PNGs kept as the fallback `<source>` so they still serve `og:image`, the tutorial page and the River page:

| | before | after |
|---|---|---|
| light | 202,631 B PNG | **45,988 B WebP** |
| dark | 130,004 B PNG | **31,346 B WebP** |

Also swapped the no-op `loading="eager"` for `fetchpriority="high"`, which is the attribute that actually affects LCP.

### Regressions this PR had introduced

- **The hero lead paragraph silently lost its styling.** It was targeted by `.home > .content > p:first-of-type`, and wrapping the hero in a grid meant no `<p>` was a direct child of `.content` any more. It was rendering `16px/24px/black` instead of `18.4px/31.28px/#374151`. Retargeted to `.hero-copy`; verified back to `18.4px / 31.28px / rgb(55,65,81)`, matching `main`.
- **Heading order became h1 → h3 → h2** once the cards moved above How Freenet Works. Cards promoted to `h2`, CSS retargeted. Now `h1` then six `h2`s, monotonic.
- **`aria-label` overrode the img alt**, so a screen reader got "More about River, link" and no description of the only product image on the page. Dropped; the chrome span is `aria-hidden="true"` and the alt becomes the accessible name.

### Correctness

- The dark screenshot is **1290×867**, a different aspect ratio from the light one's **1379×1269**, so the reserved box was the wrong shape in dark mode. Each `<source>` now carries its own dimensions. **Same defect fixed on the River page**, which is where this markup was copied from.
- The navbar logo `max-height` rule was **dead code**: it loses to an `!important` rule earlier in the file, so computed `max-height` stayed 60px, not 32px. Removed, and the comment no longer claims to shrink the logo. The Donate-button change alone fixes the 320px overflow.
- `latest-news` turned `limit="0"` into 5 via `|| 5`, contradicting the Hugo branch where 0 means show everything. Nothing hits it today, but it was a trap for the next caller.

### Contrast, measured from rendered pixels

The review correctly noted my earlier check compared the *declared* colour rather than the alpha composite. Re-measured by sampling actual pixels:

| | before | after | bar |
|---|---|---|---|
| `.cta-secondary-btn` border, light | 1.70 | **3.10** | 3.0 |
| `.cta-secondary-btn` border, dark | 2.14 | **3.72** | 3.0 |
| `.home-addr` text, light | 4.34 | **5.82** | 4.5 |

My first attempt at 0.6 alpha still composited to only ~2.6:1; 0.7 was needed.

### Layout and copy

- Hero stacks below **1024px** rather than 900px, so the 901–1023 band (container still narrow, grid already two columns, screenshot squeezed to its smallest rendering anywhere) is gone.
- Caveat says **"centralized demo"**, matching the word `/try/` itself uses, instead of softening it to "hosted". It also names the platforms, since River has no mobile app and the hero compares it to two mobile-first apps.
- "Try River now" pointed at the River marketing page, promising an action it did not deliver. Split into "Try River in your browser" (`/try/`) and "What River is" (`/river/`).
- Deleted the dead `.cta-secondary` block, whose comment about keeping the demo subordinate contradicted the rule that replaced it. That rationale now lives in the new rule's comment.

### Re-validated

320, 360, 390, 414, 430, 600, 768, 899, 900, 901, 1023, 1024, 1100, 1280, 1366, 1440, 1920px, both schemes: no horizontal scroll, exactly one `h1`, River CTA above the fold everywhere except a 320×568 viewport.

### Knowingly not fixed

- **River CTA below the fold at 320×568** (original iPhone SE). Degrades gracefully; everything fits by 320×720.
- **`.cta-note` at 4.55:1.** Passes AA, if barely.
- **News shows the release notice plus 2 posts.** `limit="3"` is 3 items total, which was the intent.
- **Pre-existing and out of scope:** `/about/news/` overflows 296px at 320px (identical on `main`, embedded iframes), `/about/faq/` has 14 `h1`s, and `bulma.min.css` is 648KB render-blocking, which is what actually gates LCP here.

[AI-assisted - Claude]

sanity:
## Re-review of the fix commit (f17b46a)

Because the fix commit changed real code, the earlier approval was stale, so I re-reviewed the delta `171004fb..6e62a455` rather than merging on a review of superseded content. All 8 claimed fixes verified as landed and correct, with independently re-measured numbers. It also caught **a regression the fix commit introduced**, now fixed.

### The regression: cropping the dark screenshot

Forcing both screenshots into a fixed `26rem` box with `object-fit: cover` assumed they share an aspect ratio. They do not: light is 1379×1269 (1.09:1), dark is 1290×867 (1.49:1). Any height tall enough for the light one crops the dark one **horizontally**:

| viewport band | shot column | dark width shown |
|---|---|---|
| 1024–1215 | 365px | **58.9%** |
| 1216–1407 | 459px | **74.1%** |
| 1408+ | 552px | 89.2% |

Below 1408px the MEMBERS panel was gone entirely and bubbles were cut mid-word. That contradicted both the alt text ("messages **and a member list**") and my own `object-position` comment, which existed to avoid precisely that.

The fixed height was never what fixed CLS; `align-items: start` was. Confirmed by re-measuring with natural aspect ratios restored: **still 0.063 in both schemes** under the same throttle, because the width/height attributes on each `<source>` reserve the correct box before the image lands. Height removed, both schemes now show the whole screenshot.

### Stack point 1023px → 1215px

Bulma's container is 960px from 1024 to 1215, so two columns through that band gave a ~365px screenshot beside a copy column nearly twice its height. One pixel narrower, at 1023, the same page showed a full-width uncropped shot. Now stacks through 1215 (screenshot at 798px there) with two columns from 1216, where the container widens to 1152 and the shot gets 459px.

### WebP regenerated at native resolution

The reviewer noted the 1104px downscale would upscale on a 2× display, since the stacked layout renders the image at up to 909 CSS px. Regenerated at native size; still well under the PNGs:

| | PNG | WebP |
|---|---|---|
| light | 202,631 B | **74,446 B** |
| dark | 130,004 B | **44,618 B** |

### Verified by the re-review

CLS under 0.1 at every width in both schemes (3 runs per cell, ±0.005). Lede computes byte-identical to `main`. One `h1` then six `h2`s, no `.home-section h3` left anywhere in source. The screenshot link's accessible name is now the img alt, computed from the accessibility tree. `<picture>` source order proven correct for all four webp/scheme combinations by swapping in an unsupported MIME type, with exactly one image request per load. Zero horizontal overflow across 8 pages × 14 widths × 2 schemes. `/about/video-talks/` and `/resources/why-freenet/` byte-identical to `main`. Contrast sampled from rendered pixels: border 3.10/3.72, `.home-addr` 5.79/6.88.

One correction to an earlier claim of mine: the `<source>` dimension fix on the River page is **markup correctness only**, not a CLS improvement. `/river/` measures 0.1703 in dark on `main`, on 171004fb and on 6e62a455 alike; that page's shift is webfont swap on its own hero, untouched by this PR.

### Still knowingly not fixed

River CTA below the fold at 320×568; `.cta-note` at 4.55:1 (passes AA, barely); news showing the release notice plus 2 posts (3 items total was the intent); and the pre-existing `/about/news/` 320px overflow, `/about/faq/`'s 14 `h1`s, and the 648KB render-blocking `bulma.min.css` that actually gates LCP here.

[AI-assisted - Claude]

PR #93, opened 2026-07-29
feat(ghostkey): fix donation funnel conversion and suppress Stripe Link

sanity:
## Problem

The Ghost Key donation flow asked people to scroll past the entire explanation before offering them any way to act, and the checkout page contradicted the product it was selling.

Measured on the live site before this change:

| | Desktop | Mobile |
|---|---|---|
| `/ghostkey/` height | 5,525px (6.1 screens) | 8,320px (9.9 screens) |
| Only CTA appears at | **95% down** | **96% down** |

Roughly 2,900 words before the first opportunity to donate. Separately:

- The nav **Donate** button, the most prominent donation path on the site, goes to `/donate/`, which never mentioned Ghost Keys.
- The Stripe checkout renders Link's inline signup asking for **email, mobile number and full name** at the moment of payment, on a page whose entire proposition is that the donation cannot be linked to the key.
- The "the Ghostkey Vault is experimental and can lose keys" warning appeared only on the success page, **after** payment.
- `/ghostkey/create/` contained the line "could provide additional benefits in the future", the one piece of copy on the page that read like a token sale.

## Approach

No dark patterns. Nothing added here manufactures urgency, hides the $1 option, or pressures the reader. The two structural moves are deliberately trust-first:

- The mid-page CTA is placed **after** "What Ghost Keys don't hide", not before, so nobody is asked to pay before seeing the limitations.
- The experimental-vault disclosure moves **before** payment. Disclosing a material limitation only after taking someone's money is the shape of a dark pattern even when the intent is honest.

Nothing is deleted from the explainer. The full text is unchanged; it stops being a toll gate.

The checkout page gained content but the payment form still starts **0.66 screens down on desktop**, the same as before, so no checkout friction was added.

### On the custom-amount idea, which I dropped

I had recommended adding a free-form amount field. Checking the API first showed that would be actively harmful:

- `handle_sign_cert.rs` computes `amount_dollars = amount_cents / 100` and calls `get_notary(amount_dollars)`, which reads a **per-amount** keypair (`notary_certificate_{amount}.pem`). An unlisted amount has no keypair, so cert signing fails **after the card is charged**. Worst possible failure: charged, no key.
- The amount is written into the certificate and is visible to any verifier, so fixed tiers are load-bearing for anonymity. A free-form figure like \$37.42 would be close to unique.

So instead of adding the field, both pages now explain why the tiers are fixed, turning something that looks like an unfinished form into the privacy property it actually is.

## Testing

Verified against the live Stripe API by loading the built site and scanning every frame for Link artifacts:

```
PRODUCTION (current): ['Email', 'Mobile number', 'Save my information', 'checkout with Link']
LOCAL (with fix)    : NONE (clean)
```

CTA positions after the change (built site, measured in Chromium):

| | Hero CTA | Mid CTA | Final CTA |
|---|---|---|---|
| Desktop | 5% down | 55% | 95% |
| Mobile | 4% down | 54% | 96% |

Also checked: `hugo` build clean, light and dark mode on both pages, desktop (1280) and mobile (390) renders, and that `/donate/` now links to `/ghostkey/`.

### Note on the Stripe fix

`linkAuthenticationElement` is not a valid key for `elements.create('payment', ...)`; the documented options are `layout`, `defaultValues`, `business`, `paymentMethodOrder`, `fields`, `readOnly`, `terms`, `wallets`, `applePay`. Unknown keys are ignored rather than rejected, so the previous code looked like it disabled Link while Link had been showing in production the whole time. Replaced with `wallets: { link: 'never' }`.

The `/donate/` section reuses `.funding-ghost-keys`, a style already in the theme (commented "Ghost keys — understated") that no content had ever used.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/content/donate/_index.md
+++ b/hugo-site/content/donate/_index.md
@@ -67,3 +67,13 @@ If you are interested in supporting Freenet at a significant level, we would be 
 <a href="https://www.paypal.com/donate?hosted_button_id=EQ9E7DPHB6ETY" class="funding-donate-button">Donate via PayPal</a>
 
 </div>
+
+<div class="funding-section funding-ghost-keys">
+
+## Donating by card
+
+If you would rather give by card, you can donate through [Ghost Keys](/ghostkey/). A Ghost Key is an
+anonymous cryptographic identity, issued with your donation, that works across Freenet apps and
+cannot be linked back to your payment. The minimum is $1.
+
+</div>
--- a/hugo-site/content/ghostkey/_index.md
+++ b/hugo-site/content/ghostkey/_index.md
@@ -10,6 +10,25 @@ You hold your Ghost Keys in the **Ghostkey Vault**, a Freenet delegate running i
 your Freenet node, and any Freenet app can ask the vault to sign with one to prove you
 hold a scarce, donation-backed identity, without ever learning who you are.
 
+<div class="gk-cta gk-cta-hero">
+<a href="/ghostkey/create/" class="funding-donate-button">Get a Ghost Key</a>
+<p class="gk-cta-note">$1 minimum. Freenet Project Inc is a 501(c)(3) nonprofit.</p>
+</div>
+
+## The short version
+
+- **What you get.** An identity that any Freenet app can verify, and that no one can
+  trace back to your payment.
+- **What it costs.** $1 or more, once. There is no subscription and nothing renews.
+- **Why a donation.** Identities that are free to create are free to farm. Paying for
+  one makes spam and Sybil attacks expensive, and it funds Freenet's development.
+- **How the anonymity works.** Your browser blinds the key before the donation server
+  ever sees it, so the server signs a key it cannot read.
+  [Details below](#how-it-works-blind-signing).
+- **What it does not do.** It does not make everything you sign unlinkable. Two messages
+  signed with the same Ghost Key can be tied to the same holder.
+  [We are specific about this](#what-ghost-keys-dont-hide).
+
 You can also read the [introductory article](/about/news/introducing-ghost-keys/) or
 [watch the interview](/about/news/ghost-keys-ian-interview/).
 
@@ -109,6 +128,11 @@ zero-knowledge proof over the certificate. It would require redesigning the sign
 format and the verifier, and is a substantial piece of work, but it is the direction
 we expect Ghost Keys to move in.
 
+<div class="gk-cta">
+<a href="/ghostkey/create/" class="funding-donate-button">Get a Ghost Key</a>
+<p class="gk-cta-note">$1 minimum. You now know both what it protects and what it doesn't.</p>
+</div>
+
 ## Using Ghost Keys from a Freenet app
 
 Once imported, your Ghost Key doesn't just sit in a file. It lives inside the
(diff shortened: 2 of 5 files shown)


sanity:
Two things worth a second opinion, both judgment calls rather than defects:

**1. The `/donate/` addition touches an earlier deliberate decision.** Commit b4406f27 ("replace donate page with structured funding page", Mar 2026) moved Ghost Keys *out* of the funding page and under a Resources dropdown, so adding a pointer back is arguably reversing that. I kept it deliberately quiet: it reuses `.funding-ghost-keys`, a style that commit added, commented "Ghost keys — understated", which no content has ever used. So the intent for an understated treatment appears to have existed and just never shipped.

The reason I think it belongs: the nav **Donate** button is the most prominent donation path on the site and it led to a page where the card-donation option was invisible. That is a funnel break rather than a de-emphasis. Happy to drop this hunk if the original call was to keep the funding page single-purpose.

**2. I removed some warmth from the checkout copy.** The old page had "Your generosity directly supports the ongoing development of Freenet, helping us build a more private, secure, and decentralized internet." The replacement is drier: "Anything above it directly funds Freenet's development." That was to make room for the disclosure and the who-sees-what breakdown without pushing the payment form down the page, and because the removed sentence was fairly generic. Easy to restore if you'd rather keep it.

**Not done, deliberately:** a custom-amount field, which was in my original recommendation. Checking `rust/api` first showed it would fail certificate signing *after* the card is charged (per-amount notary keypairs), and that fixed tiers are load-bearing for anonymity since the amount is written into the certificate. Both pages now explain this instead, and there's a comment on the radio group so the next person doesn't try it.

**Also spotted, not fixed here** (happy to file separately): `handle_sign_cert.rs` checks the `certificate_signed` metadata flag and then sets it in a separate call, so there's a TOCTOU window where two concurrent requests with the same PaymentIntent could each sign a key. Pre-existing and unrelated to this PR, and it needs a real fix rather than a drive-by, so I left it alone.

[AI-assisted - Claude]

sanity:
## Review

Four lenses, run sequentially in-session (not as independent blind subagents), each committed to in writing before starting the next. Five findings, all fixed in this branch. Treated as full tier because the change touches a live payment path.

### Lens 1 — factual-claims audit (the failure mode specific to this change)

Every privacy claim added to the copy, checked against `rust/api`. On a page whose value is being trustworthy about privacy, an overclaim is worse than the problem it replaces.

| Claim | Verdict |
|---|---|
| "Stripe never sees your Ghost Key" | ✅ True. `routes.rs:227-234` sets only `donation_type: freenet` in metadata; the blinded key goes to `/sign-cert`, never to Stripe. |
| One donation issues one key | ✅ Holds. `handle_sign_cert.rs:85` rejects a PaymentIntent already carrying `certificate_signed`. |
| Amount is per-tier and fixed | ✅ Confirmed, `delegates.rs:34-46` + `handle_sign_cert.rs:110-117`. |
| "the donation server sees that a donation happened, the amount" | ❌ **Understated.** `handle_sign_cert.rs:58` retrieves the *full* PaymentIntent and `:69` logs it at INFO. Fixed in 77d81144: the copy now says the server handles your payment record, and explains why the link still never exists. |
| "Nobody, including us, ever holds both halves" | ❌ **Flourish stronger than the project's own claim**, and ambiguous given the above. Removed in 77d81144. |

### Lens 2 — adversarial bug hunt

Anchors all resolve (`#how-it-works-blind-signing`, `#what-ghost-keys-dont-hide`), no stale copy left, light and dark both render, mobile clean.

One real finding: the donation tiers are now hardcoded in **three** places (the radio group, the main-page prose, and the per-amount keypairs on the API host). A future edit to one drifts from the others, and the failure mode is a donor charged with no key. Pinned with a comment on the radio group in 77d81144.

### Lens 3 — code-first read of the diff

That pin comment was HTML, so it shipped to every visitor. Changed to a Go-template comment in 452acd48 so it stays in source only.

### Lens 4 — big-picture / regression

**The most useful finding.** `introducing-ghost-keys.md:130` links straight to `/ghostkey/create/`, so a reader can reach the payment form without ever passing through the explainer. The checkout page discussed what stays private but never mentioned that a Ghost Key is a persistent pseudonym, and never linked to "What Ghost Keys don't hide". A reader arriving by that path could pay having seen only the upside.

Fixed in 4b302f64, placed **above** the payment form rather than below it.

### Verification

Against the live Stripe API, scanning every frame:

```
PRODUCTION (current): ['Email', 'Mobile number', 'Save my information', 'checkout with Link']
THIS BRANCH         : NONE (clean)
```

| | Before | After |
|---|---|---|
| `/ghostkey/` first CTA (desktop) | 95% down | **5% down** |
| `/ghostkey/` first CTA (mobile) | 96% down | **4% down** |
| `/ghostkey/create/` form top (desktop) | 592px | 704px (still under one screen) |

Also: `hugo` build clean, three CTAs at 5/55/95% desktop and 4/54/96% mobile, limitations link at y=345 vs form at y=704, `/donate/` links to `/ghostkey/`.

### Not fixed here

`handle_sign_cert.rs` checks `certificate_signed` and sets it in a separate call, so two concurrent requests with the same PaymentIntent could each sign a key. Pre-existing, unrelated to this PR, and wants a real fix rather than a drive-by. Say the word and I'll file it.

[AI-assisted - Claude]

PR #113, opened 2026-08-16
feat(homepage): surface Ghost Keys in the For Supporters card

sanity:
## Problem

Ghost Keys are unreachable from the homepage. The nav **Donate** button goes to `/donate/`, which links on to `/ghostkey/` since #93, but nothing on the front page itself mentions them.

The constraint when we discussed this was that the homepage is already arguably too busy, so the fix shouldn't add weight.

## Approach

It doesn't add any. The **For Supporters** card carried one link against two in each of its neighbours, with visible empty space below it:

| card | links before | links after |
|---|---|---|
| For Users | 2 | 2 |
| For Developers | 2 | 2 |
| For Supporters | **1** | 2 |

Filling that existing gap balances the row. **Rendered page height is unchanged at 1944px** on desktop.

The card's second sentence ("Your support helps build decentralized internet infrastructure that matters") was generic filler, so it makes way for a concrete one that also explains the new link. Without that, "Ghost Key" arrives as unexplained jargon on the front page.

## Testing

`hugo` build clean. Verified in Chromium at 1280 and 390 wide: exactly one `/ghostkey/` link on the page, desktop height unchanged at 1944px, and the three cards render evenly.

## Note on framing

We'd discussed a "membership" reframe and recurring billing, which would change this copy. Neither is built, so this describes what actually exists today: a one-time donation that issues a key. Worth revisiting the wording if the recurring model ever lands.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/content/_index.md
+++ b/hugo-site/content/_index.md
@@ -83,10 +83,12 @@ maintain, no cloud bills, no terms of service.
 
 ## For Supporters
 
-Freenet is built by a small team, funded through grants and donations. Your support helps build
-decentralized internet infrastructure that matters.
+Freenet is built by a small team, funded through grants and donations. Donating can also issue you a
+Ghost Key, an anonymous identity that works across Freenet apps.
 
 <a href="/donate/" class="funding-learn-more">Support Freenet →</a>
+
+<a href="/ghostkey/" class="funding-learn-more">Get a Ghost Key →</a>
 
 </div>
 


sanity:
Code-first review lens run (read the code before the description, looking for mismatches). No substantive findings. It independently verified every falsifiable claim:

- **Desktop height parity confirmed.** Built both `origin/main` and this branch, measured in Chromium at 1280px: both exactly 1944px. The reviewer identified *why*, which is worth recording: `.home-sections` is a 3-column grid with default `align-items: stretch` (`freenet.css:1031-1036`), so row height is set by the tallest card. The two sibling cards were already two links deep, so the new link is absorbed into height they had already established.
- **Mobile does grow, and this PR does not claim otherwise.** At 390px the grid collapses to one column (`freenet.css:1038-1042`), so the extra link is not absorbed: 3558px → 3625px, +67px. Recording it here so the claim isn't read as broader than it is.
- **`check-links` CI gate passes**: `scripts/check-links.py` against the built site reports no broken internal links across 222 pages.
- **goldmark rendering**: both raw `<a>` tags render as separate un-mangled `<p>` elements at the pinned Hugo version.
- **The removed sentence** appears nowhere else in the repo. No meta description, test, or snapshot depended on it (the homepage layout emits no meta description at all).
- Exactly one `/ghostkey/` link on the page.

Two non-blocking observations from the reviewer: stacked same-style links are the pre-existing pattern in the sibling cards rather than something new here, and link text is distinct so there's no ambiguous-link-name accessibility issue. The card's use of raw `<a>` rather than markdown link syntax is pre-existing inconsistency in the file, untouched by this PR.

[AI-assisted - Claude]

PR #120, opened 2026-08-25
feat: give users a signposted route to report a bug

sanity:
## Problem

An ordinary user on freenet.org who hits a bug has nowhere correct to go.
Concretely, on the live site today:

- The strings "report a bug", "file an issue" and "bug report" appear
  nowhere in the site's content or templates.
- The only GitHub link in the footer is the "src: freenet/web" credit, so
  the single most obvious place a user would click routes every River bug
  and every node bug into the *website's* tracker.
- /community/support/ and /community/get-involved/ are published but
  completely empty (title front matter, no body). Quickstart's "What's
  Next" list links to the latter, so the one navigational promise of a
  community route is a blank page.
- Nothing anywhere states which repo covers what. /apps/ links ten repos,
  but as project homepages, not as "report bugs here".

Freenet spans freenet-core, river, web, ghostkeys and the rest, so "file
an issue" is ambiguous even for someone willing to look.

## Approach

Fill the existing empty /community/support/ page with a routing page that
sorts a reporter by what they were doing, rather than adding new
scaffolding. It covers River, the node itself, the website, Ghost Keys,
other apps, and an explicit "Not sure" route (freenet-core, which the
maintainers watch most and can move a report from).

It says plainly that filing on GitHub needs a GitHub account, and offers
Matrix as the route for people who do not want one. It also documents
`freenet service report`, which produces a report code maintainers can
look up, along with `--local` for anyone who would rather inspect the
bundle before uploading it.

Discovery: one footer link, present on every page, plus an entry in the
homepage's secondary-links row and in Quickstart's Troubleshooting and
"What's Next" sections, which is where a struggling user already is.

The empty get-involved stub is deleted; /community/get-involved/ and a
new short /report/ both alias to the page, so the Quickstart link that
previously dead-ended now lands somewhere real.

## Testing

- `hugo` builds clean; `scripts/check-links.py` reports no broken
  internal links across 223 pages.
- Alias redirects verified in the built output for /report/ and
  /community/get-involved/.
- All five external GitHub targets return 200.
- Rendered and screenshotted at 1200px and 390px, light and dark
  palettes covered by the CSS.

[AI-assisted - Claude]

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,7 +6,7 @@ We welcome contributions to Freenet! Here's what you need to know.
 
 - **Claim the issue first.** Comment on the issue you want to work on and wait for a maintainer to confirm before starting work. This prevents duplicate effort and ensures the approach is aligned with project direction.
 - **Read the codebase conventions.** See [AGENTS.md](AGENTS.md) for project structure, coding standards, and testing requirements.
-- **Ask questions.** If something is unclear, ask on the issue or in our [Matrix channel](https://matrix.to/#/#freenet:matrix.org) before writing code.
+- **Ask questions.** If something is unclear, ask on the issue or in our [Matrix channel](https://matrix.to/#/#freenet-locutus:matrix.org) before writing code.
 
 ## Quality Standards
 
@@ -31,7 +31,7 @@ We require the use of a frontier-class model (as of early 2026: Claude Opus 4, G
 
 ## Getting Help
 
-- [Matrix chat](https://matrix.to/#/#freenet:matrix.org) for questions and discussion
+- [Matrix chat](https://matrix.to/#/#freenet-locutus:matrix.org) for questions and discussion
 - [Issue tracker](https://github.com/freenet/freenet-core/issues) for bugs and feature requests
 - [API docs](https://docs.rs/freenet) for code reference
 - [Freenet manual](https://freenet.org/resources/manual/) for architecture overview
--- a/hugo-site/content/_index.md
+++ b/hugo-site/content/_index.md
@@ -131,6 +131,7 @@ just a few hops, scaling efficiently to millions of peers, no servers required.
 
 [River](/river/) · [Apps](/apps/) · [Video Talks](/about/video-talks/) ·
 [Matrix Chat](https://matrix.to/#/#freenet-locutus:matrix.org) ·
-[GitHub](https://github.com/freenet/freenet-core) · [FAQ](/about/faq/)
+[GitHub](https://github.com/freenet/freenet-core) · [FAQ](/about/faq/) ·
+[Report a Bug](/community/support/)
 
 </div>
--- a/hugo-site/content/build/manual/community.md
+++ b/hugo-site/content/build/manual/community.md
@@ -7,4 +7,4 @@ aliases:
 ---
 
 - [Github](https://github.com/freenet/freenet-core)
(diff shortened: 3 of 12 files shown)


sanity:
## Review

Light-tier change (site content plus a footer link), so one independent adversarial reviewer, blind to the authoring, then a second verification pass over the fixes because the content changed materially between them. Both lenses read the checked-out branch and traced claims to primary source rather than to the page. External model pass not run; substituted a second Claude lens with a distinct brief, per `multi-model-review.md`.

Every finding was either fixed or dismissed with a stated reason. Nothing was carried forward.

### Round one

**Blocking.** The privacy disclosure for `freenet service report` understated the upload. Verified in `report.rs`: it also sends the machine's hostname (`SystemInfo.hostname`, `report.rs:169`) and the serialized `NodeDiagnosticsResponse`, which carries the peer ID, the addresses of every connected peer, and the held contracts. The page now says all of that, including that the receiving end records the sender's IP. On a site whose pitch is privacy, "logs and your config file" was the wrong summary.

**Should-fix.** "each has a `CONTRIBUTING.md`" was false (present in freenet-core, river and web only, absent in the other seven). The described order of operations was reversed: `get_user_message()` runs last, after a node query that can block. `--local` is a dead end, and the page did not say you have to re-run. Raven was missing from the routing and the other-apps route cost three hops. The prose had no contractions anywhere, which read stiff against `/try/` and `/uninstall/`. `/community/` had no `_index.md`, so Hugo titled it "Communities".

**Nits.** British "agonising"; lowercase "the vault"; a yellow warning box used for the ordinary fact that GitHub needs an account; duplicated footer-link CSS.

### Round two

**Should-fix.** The Ghost Keys entry sent donation and certificate problems to `freenet/ghostkeys`, which cannot fix them: the Stripe integration and the certificate signer are `rust/api/` in this repo. Split, since routing the reporter correctly is the one thing this page cannot get wrong. Three further links to the wrong Matrix room, including `/donate/thanks/`, the page a donor lands on right after paying.

**Nits.** The upload lists every contract the node holds, not only subscribed ones (`collect_contract_states` takes the empty-`contract_keys` branch, `p2p_protoc.rs:640`). The up-to-a-minute wait was attributed to the wrong condition: connection-refused is not retried, so a node that simply is not running fails at once. Re-running without `--local` collects a fresh bundle rather than sending the one you just read. The suggestion to send the report code over Matrix instead of a public issue was dropped: there is no public read-by-code endpoint, so the code is not a retrieval vector, and the Matrix room is public anyway, so it implied a privacy difference that does not exist.

### Dismissed, with reasons

- The inert `description:` front matter. The template emits no `<meta name="description">` and five other pages do the same, so it is a pre-existing pattern rather than a regression introduced here.
- Prettier's 4-space indent in `footer.html`, and its whole-file reformat of `freenet.css`. Both files were already prettier-dirty on `main` and no CI job checks formatting, so reformatting them here would bury a small change in churn. The new markdown is prettier-clean.

### Verification

`hugo` builds clean; `scripts/check-links.py` reports no broken internal links across 223 pages; both aliases generate redirect stubs; all sixteen external links return 200; rendered and screenshotted at 1200px and 390px in light and dark.

One thing is out of scope for this repo: `freenet-core`'s own `CONTRIBUTING.md` still links `#freenet:matrix.org`, the IRC bridge. Filed separately.

[AI-assisted - Claude]

PR #124, opened 2026-08-29
chore: remove locut.us domain from dashboard links

sanity:
## Problem
`nova.locut.us` is registered to Ian's home address and appears in public-facing pages (quickstart guide, slide-deck demo pages).

## Approach
The old `http://nova.locut.us:3133/` dashboard URL already 301-redirects to the canonical `https://telemetry.freenet.org/`, so this swaps all three references to use the canonical domain directly instead of the personal one.

## Testing
- Verified `curl -sI http://nova.locut.us:3133/` returns a 301 to `https://telemetry.freenet.org/`.
- Verified `https://telemetry.freenet.org/` resolves and returns 200.
- Not otherwise built/tested (content-only Hugo edits).

## Not touched
`hugo-site/static/keys/gateways.toml` still references `nova.locut.us` and `vega.locut.us` — that file is being handled separately in #123 and was explicitly out of scope here.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/content/quickstart/_index.md
+++ b/hugo-site/content/quickstart/_index.md
@@ -63,7 +63,7 @@ Need to remove Freenet? See the [uninstall guide](/uninstall/).
 
 ## What's Next?
 
-- [Live Network Dashboard](http://nova.locut.us:3133/) - Watch real-time activity on the network
+- [Live Network Dashboard](https://telemetry.freenet.org/) - Watch real-time activity on the network
 - [User Manual](/build/manual/) - Learn how Freenet works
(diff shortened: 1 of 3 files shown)


sanity:
Superseded by #84, which is yours and predates this by a month.

#84 covers the same three files plus `hugo-site/themes/freenet/layouts/shortcodes/network-hero.html`, which this PR missed, so it is the more complete change. The only other difference is trailing slashes on the URLs.

This was opened by an agent sweeping the personal domain out of the public repos; it did not check for an existing open PR covering the same files first. My instruction to it was at fault, not its work.

[AI-assisted - Claude]

PR #125, opened 2026-08-29
docs: stop calling the browser version a centralized demo

sanity:
## Problem

The site told visitors the browser version of River was a **centralized demo**:

> The browser version is a centralized demo running on our server.

and, on `/try/`:

> This is a hosted demo. It runs on our server…
> The hosted demo is centralized, which is the opposite of what Freenet is for.

That is misleading. `try.freenet.org` is a real Freenet peer we run, fronted by
`freenet-gateway-proxy`. A visitor there loads the real River web contract
(`raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv`) and joins the same **Freenet Official**
room that people running their own peers are in. Messages go over the real network. Nothing
about the app or the network is a mock-up, so "centralized demo" told visitors their session
was a simulation of Freenet rather than Freenet itself.

## Approach

Say the accurate, narrower thing: you are **using our peer instead of your own**, so your keys
live on our peer rather than your machine — and that is the reason to install.

Existing claims that are still true were left alone: the access-key encryption line on `/try/`,
the one-click export to your own peer, and the firewall-fallback note in the quickstart.

Files changed (copy only, no logic):

- `content/_index.md` — home hero note under the CTA row
- `content/try/_index.md` — the warning callout and the closing section ("Ready for the real thing?" → "Ready to run your own peer?")
- `content/quickstart/_index.md` — "look before installing" intro, and "hosted demo" → "browser version" in the firewall note
- `themes/freenet/layouts/shortcodes/river-invite-button.html` — the `hosted="true"` variant's note

## Testing

`hugo --minify` builds clean; new copy verified in the rendered `public/index.html` and
`public/try/index.html`. No logic touched.

[AI-assisted - Claude]

### DIFF
--- a/hugo-site/content/_index.md
+++ b/hugo-site/content/_index.md
@@ -22,9 +22,10 @@ in the middle.
   <a href="/try/" class="cta-secondary-btn">Open River in your&nbsp;browser&nbsp;→</a>
 </div>
 
-<p class="cta-note">The browser version is a centralized demo running on our server. Installing your
-own peer takes a couple of minutes on Windows, macOS or Linux, and gives you the real thing, with no
-server in the middle.</p>
+<p class="cta-note">The browser version is the real River on the real Freenet network. You're just
+borrowing a peer we run instead of one on your own machine, so your keys live with us. Installing
+your own peer takes a couple of minutes on Windows, macOS or Linux, and keeps everything on your
+side.</p>
 
 </div>
 
--- a/hugo-site/content/quickstart/_index.md
+++ b/hugo-site/content/quickstart/_index.md
@@ -10,9 +10,9 @@ It's the best way to see the network for real.
 The room runs as a Freenet contract: code that lives across the peer-to-peer network instead of on
 a central server. No company hosts it, no account to make, no admin who can shut it down.
 
-Want to look before installing? [Try River in your browser](/try/), a hosted demo of this same
-room with no download. It runs on a peer we host instead of one on your machine, so it's a
-preview rather than the real thing.
+Want to look before installing? [Try River in your browser](/try/), the same room on the same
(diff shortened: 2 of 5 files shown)


sanity:
Ran an independent accuracy review of the new copy (Light tier; copy-only diff, no logic). Verified against the live deployment rather than the docs:

- **"Real River, real network" is accurate.** `try.freenet.org` is served by `freenet-hosted.service` on nova — `freenet network --hosted-mode --network-port 31345 ...`, a genuine peer that joins the P2P network, fronted by the route-filtered `freenet-gateway-proxy`. It serves the real River contract key `raAqMhMG7KUpXBU2SxgCQ3Vh4PYjttxdSWd9ftV7RLv` (HTTP 200, real River shell), the same key used everywhere else.
- **The retained "even we can't read them" line is true.** This claim predates the PR, so it got checked hard. P1+P2 of freenet/freenet-core#4381 did land (#4620, #4723): `secrets_store/user.rs` derives a per-user DEK from a client-held bearer token via BLAKE3, independent of the node KEK, gated on `--hosted-mode` — which the nova process has set. While the token isn't presented, the server holds ciphertext it has no key for. The copy correctly scopes it to "while you're not using it" and does not claim confidentiality during an active session, which would be false (the delegate runs server-side and needs plaintext).
- **No overcorrection.** Every touchpoint keeps "your keys live with us" prominent; `/try/` leads with it in bold.
- **No broken links or shortcode issues**; no other content page still carries the old framing.

Follow-up outside this repo: `freenet-gateway-proxy-hosted/README.md` still says P1/P2 "do not exist yet" and warns against running against real user data. That was true when it was written (last commit 2026-06-19) and is now stale — both phases are live in production. Worth a doc update there.

[AI-assisted - Claude]

PR #163, opened 2026-09-02
docs: add Docker install instructions

sanity:
## Problem

[freenet-core#5516](https://github.com/freenet/freenet-core/pull/5516) is merged and publishes an official container image to `ghcr.io/freenet/freenet-core` for every release, for `linux/amd64` and `linux/arm64`. Nothing on the site mentions it.

## What this adds

A new page at `/quickstart/docker/` covering the `docker run` one-liner, a compose file, ports, how to check on a running node, and the two things that are easy to get wrong.

**It keeps itself updated.** Worth stating plainly, because the usual container habit is to pull a new image when you remember. A Freenet peer that falls far enough behind is refused by the network rather than merely missing features, so the image applies updates itself the way the desktop install does. No Watchtower, no cron job.

**It wants `--network host`.** Under Docker's default bridge network the dashboard is unreachable: the local API binds loopback, which under bridge is the *container's* loopback, and publishing the port does not help because nothing is listening on an address it forwards to. UDP hole punching also degrades, because bridge networking rewrites the source port. The bridge fallback is documented along with what it costs, for people on Docker Desktop who have no choice.

## Placement

Docker is offered **inside the Linux tab** of the install picker rather than as a fourth tab.

The tabs select an operating system and Docker is not one. Host networking, which is the mode where the dashboard actually works, is Linux in practice anyway. And someone who just wants Freenet on their own laptop should still see exactly one instruction, not a fork in the road.

It renders as a quiet footnote under the install command, not a competing option:

> Running a server, a NAS or a Raspberry Pi? There's an official container image that keeps itself updated: run Freenet in Docker.

Happy to move it if you'd rather it were more prominent, or add it to the nav.

## Also

Rewrote the "Containers & headless servers" troubleshooting note. It described the installer failing inside an existing container, which now reads as discouraging when there is a supported image. It points at the image first and keeps the `--system` advice for people running the installer in a container themselves.

## Testing

`hugo` builds clean, `/quickstart/docker/` renders, and the links resolve from both the Linux tab and the troubleshooting note.

## Do not merge yet

**The image does not exist until a release publishes it.** freenet-core#5516 only just merged, and `docker-publish.yml` fires on `release: published`, so `ghcr.io/freenet/freenet-core` has no tags right now. Merging this before then would ship instructions that fail with what looks like a missing image.

There is also a one-time step after that first publish: GitHub creates the package **private** by default, so its visibility has to be set to public once before anyone outside the org can pull it.

So the order is: release goes out, image publishes, package set public, then this merges. I am tracking the first three and will comment here when they are done.

[AI-assisted - Claude]

https://claude.ai/code/session_01W2mLM7JF3KC6UfTB4Zqwf6

### DIFF
--- a/hugo-site/content/quickstart/_index.md
+++ b/hugo-site/content/quickstart/_index.md
@@ -51,8 +51,10 @@ for help, or see [Report a Bug or Get Help](/community/support/) for where to fi
 fresh invite code. If you see the room but can't send messages, click the **"i"** icon next to the
 room name, click **"Leave Room"**, then get a new invite.
 
-**Containers & headless servers:** If service installation fails (common in LXC/Docker), use the
-system-wide service instead: `sudo freenet service install --system`
+**Containers & headless servers:** There is an official container image, and it is the easier
+route on a server: see [run Freenet in Docker](/quickstart/docker/). If you are instead running
+the installer inside an existing container or LXC and service installation fails, use the
+system-wide service: `sudo freenet service install --system`
 
 **Network requirements:** Freenet uses UDP hole punching for peer-to-peer connections. Most home
 routers support this without configuration. Strict corporate firewalls may block connections. If
--- a/hugo-site/content/quickstart/docker.md
+++ b/hugo-site/content/quickstart/docker.md
@@ -1,0 +1,138 @@
+---
+title: "Run Freenet in Docker"
+date: 2025-01-01
+draft: false
+---
+
+Freenet publishes an official container image for every release. It is built for
+`linux/amd64` and `linux/arm64`, so it runs on an ordinary server, a NAS, or a
+Raspberry Pi.
+
+This page is for running a peer on a machine you administer. If you just want to
+use Freenet on your own computer, the [normal install](/quickstart/) is simpler
+and does more for you.
+
+## Start a node
+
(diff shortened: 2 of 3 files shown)


sanity:
The blocker is cleared: the image is published and publicly pullable.

Verified unauthenticated (logged out of ghcr.io first), so this is what a stranger gets:

```
anonymous manifest fetch                          HTTP 200
docker run --platform linux/amd64 ... :v0.2.133   Freenet version: 0.2.133 (d6f099508f36)
docker run --platform linux/arm64 ... :v0.2.133   Freenet version: 0.2.133 (d6f099508f36)
docker run                        ... :latest     Freenet version: 0.2.133 (d6f099508f36)
```

The commit hash matches the v0.2.133 release commit, and the tags `latest`, `0.2`, `0.2.133` and `v0.2.133` all resolve.

So every instruction on the new page now works as written.

One note for whoever reads this: the v0.2.133 release notes lead with #5515, which repairs permanent state divergence where dropped broadcast UPDATEs were excluded from every later delta while the sender believed them delivered. Docker is the second item. This PR documents the image and does not describe the release, so nothing in it implies a Docker-only cut.

[AI-assisted - Claude]

PR #165, opened 2026-09-02
content: refresh the River hero screenshot in both schemes

sanity:
## Why

The light-mode hero was a capture whose legible content was users reporting that the tutorial was outdated and that there was no search engine, plus four consecutive GitHub-bot notifications:

> **Jonte-P:** How do I access the "search engine"? I've heard that there is one
> **Lukas 2.0:** I don't think there is a search engine right now... There wouldn't be anything to search at the moment anyway.
> **EricFala:** I'm trying to follow [the tutorial], but `fdev build` doesn't work... Perhaps it is outdated?
> **Ian Clarke:** yes, it's probably outdated

At 1379px shown in a 552px slot that text rendered around 5px, so nobody read it at a glance — but it was there for anyone who zoomed or opened `/images/river-hero-light.webp` directly.

## Real capture, not a mock-up

Recaptured from a **live River against Ian's own peer**, so this is a real room with real people.

Synthetic chat content was considered and rejected. The image is labelled "a live River chat room", so inventing messages and attributing them to a real room with named members would be a fabricated record. It is also self-defeating: invented chat reads as fake almost immediately, and the texture that makes a real capture convincing (someone saying "Hiiiii", a join notice mid-thread) is exactly what a mock-up loses.

## Two things changed beyond the content

**Capture scale.** The hero slot is 552px wide. The old 1379px capture was downscaled to 0.40, which is why its text was illegible. Captured at a 900px viewport instead, the scale is 0.579 and the conversation actually reads at normal viewing size. River keeps its three-pane layout at that width.

**Matched aspect ratios.** Both schemes were captured in one pass at the same viewport and scroll position, so both are now **1800x910**. They were previously 1379x1269 and 1290x867 — a mismatch that was the documented reason `.home-browser` could not use a fixed height. That comment has been updated rather than left describing a constraint that no longer applies.

## Privacy

The room list is cropped to the seven rooms Ian chose to show. `Politics`, `Ian / Steven`, the Direct Messages list and `Archived` are out of frame. Note `Freenet Devs` carries a padlock and **is** in frame — that is deliberate, not an oversight; an earlier version of this description implied nothing private was visible, which was imprecise.

## Verification

| | before (production control) | after |
|---|---|---|
| CLS light | 0.0142 | 0.0154 |
| CLS dark | 0.0154 | 0.0154 |
| displayed | 521x479 light / 521x350 dark | 521x263 both |

No regression, and the two schemes now match. Full sweep over home, quickstart, river, faq, news, donate and ghostkey in both schemes plus mobile: no page errors, no transparent text, no horizontal overflow, and the same **7 pre-existing `/river/` contrast failures and nothing new**.

[AI-assisted - Claude]

---

### Update

The first push of this PR had a worse problem than the image it replaced: its top message was unattributed gossip about a real person, made more legible by the change. Recaptured, and the other three consumers of these files (which I had not checked) are fixed. See the [review-response comment](https://github.com/freenet/web/pull/165#issuecomment-5518118200).


sanity:
## Review findings addressed (4446709a)

Two of these were serious, and the first one means the PR as originally pushed was worse than what it replaced.

### The replacement screenshot defeated its own purpose

The topmost message read **"unpopular and I think he got upset"** — an unattributed fragment about an unnamed third party, author line cropped off.

The whole argument for this PR was that the old capture's embarrassing content was legible if you zoomed. The replacement made text *deliberately* more legible and landed the payoff on a remark about a real person being unpopular, readable with no zoom at all — and, via `og:image`, in every social share card for `/river/`. I missed it because I was checking for complaints about the docs rather than reading the top line on its own terms.

Recaptured on the Delta thread: a join notice, a real question ("if i publish a page on Delta, will it still accessible by others when my node isn't running?"), the answer ("Yes"), and "yay". The frame now starts on a complete element instead of slicing a bubble, so there is no orphan fragment at either edge.

### /river/ was left declaring the old dimensions

`themes/freenet/layouts/river/list.html` still said 1290x867 / 1379x1269 for files this PR resized to 1800x910 — reserving a box 200px taller than the image. I verified the homepage only.

| | reserved | settled | collapse |
|---|---|---|---|
| before fix, light | 494px | 294px | **−200px** |
| after fix, light | 294px | 294px | 0 |
| after fix, dark | 294px | 294px | 0 |

Measured CLS on `/river/`: **0.164 light / 0.169 dark**, against production's 0.178 / 0.173. So it now sits slightly *below* the pre-existing baseline rather than the 0.202 the broken state produced. (The ~0.17 baseline is font-driven and pre-existing, not from this PR.)

Worth noting the reviewer's point that "I measured CLS and it didn't change" was not sufficient here — the homepage figure was genuinely unchanged while a different page regressed.

### Also fixed

- **Stale alt text** on `river/list.html` still said "Freenet Official"; the room is named "Freenet". I had corrected the identical string in `_index.md` and missed this one.
- **`static/slides/demo/river.html`** had `max-height` but no `max-width`, and the intrinsic width grew 1290 → 1800. Added `max-width: 100%`.
- The comment explaining the split dimensions is updated, since the two screenshots now share one ratio.

### Corrections to the PR description

- **Scale figure.** I wrote 0.61; measured it is **0.579** (520.9px slot / 900px capture). The 0.40 → 0.58 improvement stands.
- **"Private entries are out of frame" was imprecise.** `Freenet Devs` carries a padlock and *is* in frame. That was a deliberate choice — Ian picked the framing that ends the list around "Freenet Builders / Self Hosting" — but the wording implied nothing private was visible, which is not accurate. `Politics`, `Ian / Steven`, the DM list and `Archived` are the entries that are out of frame, and the reviewer confirmed none of those strings appear in any of the four files.

### Knowingly not changed

- **Hero balance at 1024** (~370px empty column). Real, but `align-items: start` is load-bearing: the CSS comment records that centering measured CLS 0.12 from webfont reflow. Rebalancing properly means revisiting the grid, which is a separate change from swapping a screenshot.
- **No `srcset`/`sizes`.** A 1800px file serves a ~521px slot. Only ~70KB, and adding responsive sources is orthogonal to this PR.
- **The scroll-to-bottom chevron** is a real UI affordance in a real capture; removing it would mean editing the screenshot.

[AI-assisted - Claude]

sanity:
## Re-review findings addressed (d1557fc0)

The re-review confirmed the recapture is clean — no gossip, no private rooms, no complaints, and light/dark parity verified structurally via a text-mask XOR rather than by eye. It then raised a better objection about the content, which I've acted on.

### The featured exchange was making a promise the project can't currently keep

> "if i publish a page on Delta, will it still accessible by others when my node isn't running?" → **"Yes"**

That would have been the single most legible line on the front page, and in the `/river/` `og:image` unfurl — an unqualified durability guarantee about precisely the property that is the hardest open problem right now.

The rest of the site is careful here: the hero note says "No mobile app yet", `/river/` carries a column of alpha caveats, and the project deliberately keeps "hosting" and "storing" distinct. A claim embedded in a PNG can't be softened by a prose edit later; it needs a recapture.

Recaptured on an exchange that shows the product without asserting anything: community banter ("Does macos count?"), a greeting, and a developer asking about building a TypeScript dApp on Freenet Skills. That last one is arguably better signal than the answer it replaces — it shows someone building.

### Also fixed

- **`/river/` was serving the PNG as its LCP** with `fetchpriority="high"` while an identical WebP sat in the repo. Dark was 130KB on production and had regressed to 192KB on this branch; it now serves the **83KB WebP** on both pages. Verified via `currentSrc` and `content-length`.
- **24px dead band** below the input bar in all four images, so the declared box was 24px taller than the capture. Cropped to **1800x886**, with all five declaration sites updated together.
- **`og:image` had no dimensions.** Now declared — more relevant than before since the aspect went from 1.087 to 2.03.

### Correcting my own CLS claim

The reviewer is right and I overstated this. The `/river/` residual is **font-swap, not the image** — blocking Google Fonts drives it to 0.000 on both this build and production. Reserved == settled on production too, so the only genuine collapse was the one **this branch introduced** and has now fixed. The 0.178 → 0.164 movement is incidental to the hero being shorter, not a fix, and `/river/` still sits above the 0.1 threshold for reasons this PR does not touch.

The dimension change remains correct, and carries a win I hadn't claimed: the `` fallback previously declared 1379x1269 while the dark `<source>` said 1290x867, so any engine ignoring dimensions on `<source>` reserved a badly wrong box in dark mode. Both are now consistent.

Measured after the fix: reserved == settled == **287px** on `/river/`, 0px collapse in both schemes; homepage CLS **0.0154** unchanged.

### Raised, not fixed here

- **`static/slides/futo/river-screenshot.png`** is live and carries a `DIRECT MESSAGES` panel naming a contributor, a topic reading "Atlas (search engine)", and a pinned message calling the browser version "centralized and thus somewhat misses the point of Freenet" — which contradicts merged #125. Not a consumer of these four files, so outside this diff, but it is the same class of problem still shipping. Flagged to Ian separately rather than filed publicly, since it involves a third party's message list.
- **`app-screenshot.html` emits no `width`/`height`**, so `/build/manual/tutorial/` still shifts. Pre-existing, affects other images too, and belongs in its own change.
- **Hero balance at 1024** and **no `srcset`/`sizes`** — as in the previous comment, both are real but orthogonal to swapping a screenshot.

[AI-assisted - Claude]

